PR #6, opened 2014-06-12
Now connects to a given address string instead of a port number.

psychoticmeow:
Literally the least work possible to get this functional, no parsing of the address string to try and guess what scheme the user actually wants (`tcp`, `udp`, `unix`...), so if someone uses `https`, it will fail. Intelligent address handling can be done somewhere else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,14 @@ Here is a server that closes the connection if you send it anything.
             $conn->close();
         });
     });
-    $socket->listen(1337);
+    $socket->listen('tcp://127.0.0.1:1337');
 
     $loop->run();
-```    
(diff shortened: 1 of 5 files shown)


clue:
:+1: on the simplified API!

IMHO it makes sense to specify the listening address as a single argument.

Passing only a port number has always been ambiguous with regards to which interface we're binding to. Though for BC we *might want to consider* adding a fallback where we automatically prefix "localhost" if only a port number has been given.

For the reference: UDP/UDG sockets are out of scope for this project, as they are based on datagrams instead of streams of bytes. They're already part of [React's datagram component](https://github.com/reactphp/datagram).

PR #8, opened 2014-06-20
Split server functionality out into a trait.

psychoticmeow:
One point of contention: I did not want to define the constructor in the trait, so I added the `setLoop` function to the trait. Not sure if that is the best solution.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -2,70 +2,17 @@
 
 namespace React\Socket;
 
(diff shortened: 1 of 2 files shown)


clue:
Ping @psychoticmeow, could you describe the motivation for this change?

clue:
Closing this due to a lack of feedback. Please come back with more details if you think this is still relevant and we can reopen this :+1:

PR #10, opened 2014-09-18
Allow setting socket/tcp options for server sockets.

timgws:
This patch will enable users to set options such as `SO_REUSEADDR` and `TCP_NODELAY` on new sockets that are created before being passed to the connection emitter.

nodejs by default sets `TCP_NODELAY` and `SO_REUSEADDR` to true. I wanted an easy way to replicate this behaviour in my PHP application, so I added the setTCPOption and setSocketOption functions to enable these options to be set.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -49,7 +49,7 @@ class Server extends EventEmitter implements ServerInterface
         // apply any socket options on the connection!
         foreach ($this->options as $socket_level => $options) {
             foreach($options as $option_name => $option_value) {
-                socket_set_option($socket, $option_level, $option_name, $option_value);
+                socket_set_option($socket, $socket_level, $option_name, $option_value);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Server.php
```diff
@@ -68,4 +76,24 @@ public function createConnection($socket)
     {
         return new Connection($socket, $this->loop);
     }
+
+    public function setOption($option_level, $option_name, $option_value)
+    {
+        if (!is_array($this->options[$option_level]))
```
cboden: Wrap in parenthesis please

cboden:
I like the API and the feature.

One more thing I'd like to see is error handling. `socket_set_option` returns false if the option can't be applied. Could you check the return type and emit and error if that happens?

clue:
Thanks for the PR @timgws and the elaborate README!

I'm a bit surprised this appears to use `socket_set_option()` (from `ext-sockets`) on stream socket resources. Could you provide some info on which platforms has this been tested on?

barrylb:
I think you'll find it is necessary to do:
$real_socket = socket_import_stream($socket);
socket_set_option($real_socket, ...);
(on Windows at least you can't just call socket_set_option on a stream resource)

clue:
Thanks for your effort and filing this PR, but unfortunately we've never been able to get this to work properly.

I believe some of this may have been resolved via #64 in the meantime.

As such, I'm closing this due to a lack of feedback for now. If you feel this is still relevant, please comment on this and we can reopen this :+1:

PR #14, opened 2015-02-02
Added getLoop method as protected

JordanRL:
I recently ran into a problem where I wanted to extend the React\Socket\Connection class into my own Connection class, This meant also extending the server in order to replace the createConnection() method, however Connection objects require the event loop to be injected, and as the $loop variable is private, nothing extending the class can access it.

Instead of making the $loop variable protected, I opted to create a simple getter for $loop that can be used in extension.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -56,6 +56,11 @@ class Server extends EventEmitter implements ServerInterface
 
         return (int) substr(strrchr($name, ':'), 1);
     }
+    
+    protected function getLoop()
+    {
+        return $this->loop;
+    }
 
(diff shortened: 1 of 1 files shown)


staabm:
You can catch the loop in your construct() and pass it then to parent::__construct()

JordanRL:
Yes, but this is beside the fact that the Server class is fundamentally not built in a way that is open to extension.

cboden:
I agree with @staabm. It's a much cleaner approach and accomplishes what you need; that by definition means it's open to extension.

PR #17, opened 2015-06-02
Add support for Unix domain sockets (UDS)

RafaelKa:
This change adds support for unix domain socket transport for using unix:// for connections.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -24,14 +24,12 @@ class Server extends EventEmitter implements ServerInterface
         } elseif (filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
             // enclose IPv6 addresses in square brackets before appending port
             $localSocket = 'tcp://[' . $host . ']:' . $port;
-        } elseif (preg_match('#^(unix|udg)://#', $host)) {
+        } elseif (preg_match('#^unix://#', $host)) {
             $localSocket = $host;
-        }
-
-        if (empty($localSocket)) {
+        } else {
             throw new \UnexpectedValueException(
                 '"' . $host . '" does not match to a set of supported transports. ' .
-                'Supported transports are: IPv4, IPv6, unix:// and udg:// .'
+                'Supported transports are: IPv4, IPv6 and unix:// .'
                 , 1433253311);
         }
 


src/Server.php
```diff
@@ -18,14 +18,26 @@ public function __construct(LoopInterface $loop)
 
     public function listen($port, $host = '127.0.0.1')
     {
-        if (strpos($host, ':') !== false) {
+        $localSocket = '';
+        if (filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
```
staabm: composer.json should reflect that react requires ext-filter

RafaelKa: @staabm  
This ext. is since PHP 5.2 enabled by default. http://php.net/manual/en/filter.installation.php
I see many methods from multiple PHP ext.s used in react and no entry in composer.json for them.
What is the difference by ext-filter?

cboden: Which extensions does React use that aren't declared?

RafaelKa: @cboden f.x. PCRE (for preg_replace) is like Filter enabled per default to.

Is it maybe better to preg_match both instead of using  filter?

staabm: > This ext. is since PHP 5.2 enabled by default

you neverthelesss should declar it, because a lot of people - especially on production servers - use non-default builds (`` --disable-all`) which are stripped down to the very least for performance and security reasons.

> Which extensions does React use that aren't declared?

from within this code-window I would say e.g. `ext-pcre`

src/Server.php
```diff
@@ -18,14 +18,26 @@ public function __construct(LoopInterface $loop)
...
+            $localSocket = 'tcp://' . $host . ':' . $port;
+        } elseif (filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
             // enclose IPv6 addresses in square brackets before appending port
-            $host = '[' . $host . ']';
+            $localSocket = 'tcp://[' . $host . ']:' . $port;
+        } elseif (preg_match('#^(unix|udg)://#', $host)) {
```
clue: UDG sockets are based on datagrams instead of streams of bytes. They're already part of [React's datagram component](https://github.com/reactphp/datagram).

src/Server.php
```diff
@@ -18,14 +18,26 @@ public function __construct(LoopInterface $loop)
...
+            $localSocket = $host;
         }
 
-        $this->master = @stream_socket_server("tcp://$host:$port", $errno, $errstr);
+        if (empty($localSocket)) {
```
clue: This check can probably be avoided by moving this to the `else` branch above?

clue:
:+1: on adding support for Unix domain sockets!

However, I'm not particularly happy with the current interface:

```php
$socket->listen(0, 'unix:///var/tmp/demo.sock');
```

See #6 for an alternative implementation (with a cleaner API IMHO), which involves a BC break.

RafaelKa:
@clue 
i thought to for this solution, but i think this is not courteous for community. To many people are using this library, therefore i did this backward compatible patch. 
However i would prefer #6 at all ;).

RafaelKa:
"For UNIX domain sockets, file and directory permissions restrict which processes on the host can open the file, and thus communicate with the server. Therefore, UNIX domain sockets provide an advantage over Internet sockets (to which anyone can connect, unless extra authentication logic is implemented)."

What do you think about setting permissions on uds file at creation time?

1: Is current pull request good as is. Make new pull request after this one is merged
2: push https://gist.github.com/RafaelKa/33856a11023455653af8#file-server-php-L19-L38 in this pull request immediately

src/Server.php
clue: > you neverthelesss should declar it

I agree, *all* dependencies should explicitly be defined (and kept to a minimum for obvious reasons). Though in this case, have you checked PHP's installation instructions? (http://php.net/manual/en/pcre.installation.php and http://php.net/manual/en/filter.installation.php)

> Which extensions does React use that aren't declared?

None that I'm aware of. It anybody happens to spot one, please file an issue or even better a PR in the relevant component :-)

clue:
> What do you think about setting permissions on uds file at creation time?

I think it makes sense to have this feature. However, I'm not a big fan of the API, so perhaps we can postpone this until we find a better API? The `Server` implements the `ServerInterface`, and adding UDS parameters to the global `ServerInterface` sounds odd IMHO.

RafaelKa:
@clue, what do you mean with better API? 
is WIP for new API? 
becomes this lib new structure?

RafaelKa:
You added [support for Unix domain sockets (UDS) ](https://github.com/reactphp/socket-client/commit/2536462fb140082cea149d6f0afa922bb8defd91), i think currently is the time to go with UDS on serverside!

Are in this PR some deficits present, which need to be fixed? 
Or is this PR for garbage and i can forget about it and try something other again without new dependencies?

mikegioia:
@RafaelKa @cboden Is there anything holding up the merge of this PR? I'd love to be able to use this without forking :D :pray:

RafaelKa:
i can decide here nothing :-P only cboden can do that, not sure but i think it is on hold, because new dependency(filter) is here. If it is so, we must do the same with regex instead of filter.

cboden:
I think I would prefer a separate interface as opposed to string parsing. The string parsing is at a higher level of the chain (conceptually) in that's it's a URI whereas this lib is about socket (lower level) I/O. 

I think a helper function somewhere (wrapper lib maybe) would be a better place to do string parsing and call the correlating API of TCP or Unix socket. 

Thoughts @clue?

mikegioia:
Is it possible to mimic behaviour from [Socket-Connection](https://github.com/reactphp/socket-client/blob/master/src/UnixConnector.php) with, say, a `UnixServer` class (leaving the `Server` class alone)?

```php
$loop = React\EventLoop\Factory::create();
$socket = new React\Socket\UnixServer( $loop );
$socket->on( 'connection', function ($conn) { ... } );
```

marcj:
Can this PR merged?

clue:
> Can this PR be merged? What's missing?

Several issues have been raised in this ticket, besides it currently has a merge conflict.

> Is it possible to mimic behaviour from Socket-Connection with, say, a UnixServer class (leaving the Server class alone)?

I think this makes sense :+1: 

> However, I'm not particularly happy with the current interface:
> 
> $socket->listen(0, 'unix:///var/tmp/demo.sock');

I still think this doesn't make much sense. Ticket #19 suggests an alternative API.

clue:
I'm closing this for now as it hasn't received any input in a while and there is now a new PR #120 to implement this by @andig. Thank you for your effort and filing this PR! If you feel this PR has been closed prematurely and this problem persists, please come back with more details and we can reopen this :+1:

PR #20, opened 2015-07-04
Annotations fix

tkleinhakisa:
Removed the use of the special `@event` annotation to ease the use of the lib with annotation library, in particular Doctrine annotations

see https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/AnnotationReader.php#L55 for the list of ignored annotation

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -5,7 +5,7 @@ namespace React\Socket;
 use Evenement\EventEmitter;
 use React\EventLoop\LoopInterface;
 
-/** @event connection */
+/** Emits the connection event */
 class Server extends EventEmitter implements ServerInterface
 {
     public $master;
--- a/src/ServerInterface.php
+++ b/src/ServerInterface.php
@@ -4,7 +4,7 @@ namespace React\Socket;
 
 use Evenement\EventEmitterInterface;
 
-/** @event connection */
+/** Emits the connection event */
 interface ServerInterface extends EventEmitterInterface
 {
     public function listen($port, $host = '127.0.0.1');


clue:
> Removed the use of the special @event annotation to ease the use of the lib with annotation library

I'm not sure I follow this… :)
@tkleinhakisa perhaps you can elaborate on the motivation for this change?

A similar notation is used in quite a few of React's components. Do we have a better alternative to document the expected events or do we want to stick to plain text only?

tkleinhakisa:
Hi @clue 

In my case, i am working with the doctrine annotation library which is widely used, see:
https://libraries.io/packagist/doctrine%2Fannotations/dependents

This library defines a base set of classic/standard/allowed annotations mostly based on the phpdocumentator project

Custom annotations must be defined in a class and imported with a use statement otherwise you get an exception:

```php
use Symfony\Component\Validator\Constraints as Assert;
use Doctrine\ORM\Mapping as ORM;
/**
 * @ORM\Column(type="string")
 * @Assert\NotBlank()
 * @var string
 */
```

The addition of the `@event` annotation make it impossible to work with those classes

In my use case i wrote a symfony2 compiler pass to be able to define a ratchet websocket server as a service which ends looking like that in the config:
```
realtime.server.io_server:
        class: Ratchet\Server\IoServer
        arguments:
            - ~
            - @realtime.server.socket
            - @realtime.server.loop
        tags:
            - { name: realtime.server, priority: 0 }

    realtime.server.http_server:
        class: Ratchet\Http\HttpServer
        arguments:
            - ~
        tags:
            - { name: realtime.server, priority: 10 }
```

Every classes used in the config is parsed to read their annotations when the cache is generated and that's when i get the errors

If react's annotations are only provided as documentation, i think plain text is enough
If react's annotations are provided to enable some kind of features, then creating annotation classes would be the safest way to go

PR #27, opened 2015-10-14
Change $loop visibilty to allow inheritance use

Maxooo:
It should be nice to have access to the loop variable in a child class that extends the Server one.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -9,7 +9,7 @@ use React\EventLoop\LoopInterface;
 class Server extends EventEmitter implements ServerInterface
 {
     public $master;
-    private $loop;
+    protected $loop;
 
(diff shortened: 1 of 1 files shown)


cboden:
This can be achieved via constructor overloading without having to change the React code base:
```php
class MyServer extends \React\Socket\Server {
    private $loop;

    public function __construct(\React\EventLoop\LoopInterface $loop) {
        $this->loop = $loop;
        parent::__construct($loop);
    }
}
```

Maxooo:
First of all thank you for your quick answer. This is exactly what I''ve done, but it looks more like a workaround than a real solution. I mean, "protected" visibility exists to avoid this kind of practice and make life and code easier =)

PR #34, opened 2016-04-04
Add a more generic bind() method to allow Unix socket server

rcambien:
I also need to implements an unix socket server ( like #17 and #6 ) so I added a simpler method bind($address) to bind a server to any socket type.
I converted the listen() method to use this new method for BC.


umbri:
+1

clue:
Closed via #61

Issue #24, opened 2015-09-18
Async SSL/TLS server


clue:

clue:
A possible alternative is to launch a SSL/TLS proxy server (such as stunnel, haproxy or nginx) in front of your unencrypted socket server.

CriseX:
I'm curious why there is no 1st party support for this yet? That old pull request basically had the gist of it all right. As for the alternative, once #25 is resolved, this issue will be a little more than a matter of parity between the server and client side capabilities of react and their completeness since we can actually set up a proper IPC socket between the two at that point, rather than using the loop back interface.

Ironically, php's openssl extension, and subsequently anything TLS related, seems to suffer a bit from the same problem as openssl itself when it comes to how they are documented (or lack thereof in some cases). The PHP 5.6 migration guide is the only official documentation that mentions most of the "recent" and relevant changes to PHP as far as I know.

Personally, I prefer this https://wiki.php.net/rfc/improved-tls-defaults to the migration guide, at the time of writing this though, because it seems to still be at least as accurate and more verbose than the former.

clue:
Just a heads up: I've started looking into this while trying to add test cases for our socket-client component: https://github.com/reactphp/socket-client/pull/62

The resulting SSL/TLS socket server works reasonably well and I'll look into filing a PR for this in the upcoming weeks :shipit: 

However, we currently have to share some classes with the socket-client component, so I'll look into moving these elsewhere first. I'll keep this ticket updated.

vanderlee:
Any E.T.A. on this? Need to know if I should continue using my own hack for now or wait for this in order to meet my own deadlines.

clue:
> Any E.T.A. on this?

Unfortunately not. If your "own hack" works fine, I recommend you stick with it for the time being :+1: 

That being said, this task is high on my todo list and I plan on picking this up in the next week. However, this currently depends on https://github.com/reactphp/stream/pull/42 first if we want to avoid some rather nasty hacks here.

clue:
Progress update: All referenced issues have been resolved in the meantime. The next (last!) batch of dependencies is #49, #50 and #51.

I'm ready to file a PR for my SSL/TLS server implementation once these are finally in :shipit:

PR #35, opened 2016-04-05
Server with Transport Layer Security.

boite:
For #24.

I realise there is ongoing work in this area, but I haven't seen any concrete code that's recent. So here is a contribution to that end.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -10,25 +10,35 @@ class Server extends EventEmitter implements ServerInterface
 {
     public $master;
     private $loop;
+    private $isSecure = false;
 
     public function __construct(LoopInterface $loop)
     {
         $this->loop = $loop;
     }
 
-    public function listen($port, $host = '127.0.0.1')
+    public function listen($port, $host = '127.0.0.1', $streamContext = array())
     {
         if (strpos($host, ':') !== false) {
             // enclose IPv6 addresses in square brackets before appending port
             $host = '[' . $host . ']';
         }
+        if (isset($streamContext['ssl'])) {
+            $this->isSecure = true;
+        }
 
-        $this->master = @stream_socket_server("tcp://$host:$port", $errno, $errstr);
+        $this->master = @stream_socket_server(
+            "tcp://$host:$port",
+            $errno,
+            $errstr,
+            STREAM_SERVER_BIND | STREAM_SERVER_LISTEN,
+            stream_context_create($streamContext)
+        );
         if (false === $this->master) {
             $message = "Could not bind to tcp://$host:$port: $errstr";
             throw new ConnectionException($message, $errno);
         }
-        stream_set_blocking($this->master, 0);
+        stream_set_blocking($this->master, (int) $this->isSecure);
 
         $that = $this;
 
@@ -45,6 +55,10 @@ class Server extends EventEmitter implements ServerInterface
 
     public function handleConnection($socket)
     {
+        if ($this->isSecure) {
+            $this->enableConnectionSecurity($socket);
+        }
+
         stream_set_blocking($socket, 0);
 
         $client = $this->createConnection($socket);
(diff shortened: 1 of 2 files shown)


src/Server.php
```diff
@@ -10,25 +10,35 @@ class Server extends EventEmitter implements ServerInterface
...
+            $errstr,
+            STREAM_SERVER_BIND | STREAM_SERVER_LISTEN,
+            stream_context_create($streamContext)
+        );
         if (false === $this->master) {
             $message = "Could not bind to tcp://$host:$port: $errstr";
             throw new ConnectionException($message, $errno);
         }
-        stream_set_blocking($this->master, 0);
+        stream_set_blocking($this->master, (int) $this->isSecure);
```
boite: I would love to know if anybody has managed to solve the requirement for this master socket to block for secure connections.

CriseX: There is a way to do it... I don't have react specific code samples for it, however, the gist of it is to try until `stream_socket_enable_crypto` returns success.

```
		$result = stream_socket_enable_crypto($this->stream, true, $crypto_method);
		if ($result === 0)
			return true;

		if ($result === false)
			return false;

		$this->secure = self::STATE_ENCRYPTED;
		return true;
```

Assume the above to be part of a member function which you keep polling until the value of secure == STATE_ENCRYPTED. You don't need to do blocking at all in this way, see documentation of `stream_socket_enable_crypto`. Depending on the `stream_select` timeout values typically doing it this way results in 1-3 cycles of polling to enable cryptography (tested on loopback interface).

boite: Thank you @CriseX, your comment was very helpful. I've added the suggested improvements to this PR.

src/Server.php
```diff
@@ -16,14 +16,26 @@ public function __construct(LoopInterface $loop)
         $this->loop = $loop;
     }
 
-    public function listen($port, $host = '127.0.0.1')
+    public function listen($port, $host = '127.0.0.1', $streamContext = array())
     {
+        if (isset($streamContext['ssl']) && PHP_VERSION_ID < 50600) {
+            throw new \RuntimeException(
+                'Secure connections are not available before PHP 5.6.0'
```
boite: This is not strictly true, but I haven't been able to get it working with PHP 5.5.x using a simple HTTP server (and select loop) and web browser.  The problem seems to be that select() will permit a read of the first character (The 'G' in 'GET / HTTP/1.1') and then no more; so the http server doesn't send a response.  Works fine with wget though.  More work needed to work out what's going wrong.

src/SecureConnection.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $data = fread($stream, $this->bufferSize);
+
+        if ('' !== $data && false !== $data) {
+            $this->emit('data', array($data, $this));
+        }
+
+        if (false === $data || !is_resource($stream) || feof($stream)) {
+            $this->end();
+        }
```
boite: Note that the first read after enabling crypto seems always to be empty: so ending on an empty read will prevent a subsequent non-empty read.

profbiss:
Merge please

mpyw:
Fixed some bugs: https://github.com/boite/socket/pull/1

mpyw:
Some test cases should be added, but it deserves to be merged!

clue:
> I realise there is ongoing work in this area, but I haven't seen any concrete code that's recent. So here is a contribution to that end.

Thanks for your hard work and filling that void :+1: 

If you've kept track of #24 you'll have seen that I've kept working on resolving some outstanding design issues in the meantime, before being able to file my TLS server just yesterday :shipit:

Most of my work for this actually went into the design, testing and documentation of this feature, whereas the implementation took mere hours. Likely because these outstanding design issues have been resolved in the meantime, I've also not encountered and of the issues you've observed.

I very much value your contribution and would like to ask you to check out and/or comment on my PR so we can get any of these PRs in :+1: Happy to hear what you think about this :+1:

PR #48, opened 2016-07-29
Add sockets extension to dependencies

Deltachaos:
Not all builds of PHP contain the socket extension.


clue:
What is the motivation for this change? This component does not build on top of `ext-sockets`, but on PHP's `stream` which can not be disabled.

Deltachaos:
@clue sorry. i was a bit confused.

PR #54, opened 2016-12-29
Add functional integration tests

clue:
This simple PR aims to add a functional integration test (which also happens to be a prerequisite for implementing #24).

Unfortunately, this simple change introduces an indirect cyclic dependency between `socket -> socket-client -> dns -> socket` which causes the test setup to fail. I'm opening this PR in the hope to propose and discuss a possible solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ and [`Stream`](https://github.com/reactphp/stream) components.
   * [ConnectionInterface](#connectioninterface)
     * [getRemoteAddress()](#getremoteaddress)
 * [Install](#install)
+* [Tests](#tests)
 * [License](#license)
 
 ## Quickstart example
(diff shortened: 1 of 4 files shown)


clue:
Here's a dependency graph that shows this cyclic dependency (generated via https://github.com/clue/graph-composer):
socket


The `react/dns` package currently has a dependency on this root package (`react/socket`) and thus has no way of knowing which version we're currently at.

Note that this only applies when this is the root package, i.e. it *does* happen in a local test setup but will *not* happen when you use this project as a library and add any of these packages as a normal dependency.

One possible way to avoid this is to explicitly define the root package version as per https://getcomposer.org/doc/03-cli.md#composer-root-version:

```bash
$ COMPOSER_ROOT_VERSION=`git describe --abbrev=0` composer update
```

clue:
PR updated and ready for review :shipit:

README.md
```diff
@@ -144,6 +145,23 @@ $ composer require react/socket:^0.4.4
 
 More details about version upgrades can be found in the [CHANGELOG](CHANGELOG.md).
 
+## Tests
+
+The run the test suite, you first need to clone this repo and then install all
```
jsor: Typo: "To run..."

clue:
Thanks for spotting! Amended the last commit :shipit:

PR #61, opened 2017-01-17
Replace listen() call with URIs passed to constructor

### LINKED ISSUE #19: Clean up temporal dependence
The current socket API exhibits a [temporal dependence](http://blog.ploeh.dk/2011/05/24/DesignSmellTemporalCoupling/):

```
$server = new Server($loop);
$server->listen(1234);
$server->getPort();
$server->on('connection', function () { });
$server->shutdown();
```
- What happens if you call `listen()` twice?
  It's tempting to think that this will allow you to listen on two addresses – it does not.
- What happens if you call `shutdown()` twice or before calling `listen()`?
  Errors out – should probably be ignored
- What happens if you call `getPort()` before calling `listen()` or after calling `shutdown()`?
  Errors out – should probably be ignored
- What happens if you call `listen()` after calling `shutdown()`?
  Should work, but makes for an awkward API.

Afaict most of this boils down to this:
- What does a `Server` instance actually represent?
  Honestly, I'm not sure how to answer this. Once `listen()` has been called, it represents a "server socket", but until then?

Afaict most of these issues can be avoided if a `Server` instance _always_ were to represent a "server socket" (which is possibly in a closed state). This means that the `listen` call would have to be part of the construction.

One possible API _could_ look like this:

```
$server = $api->createServer(1234);
$server->getPort();
$server->on('connection', function() { });
$server->shutdown();
```

clue:
The current socket API exhibits a temporal dependence, i.e. it's not safe to call `listen()` or `shutdown()` multiple times on the same instance and call order is actually significant. (see #19)

This PR replaces the `listen($port, $host)` API with URIs passed to the constructor.
This means the Server now always represents a listening Socket (which may
possibly be in a closed state).
* URIs are literally the only way to provide a consistent addressing scheme throughout React's components, see reactphp/react#199
* Not all protocols know the concept of a "port", for example the future UnixServer currently would have an unused $port parameter (#17)
* The `Server` class accepts a single port parameter instead of a full URI and assumes a localhost socket in this case for better compatibility with previous releases.

This is obviously a BC break from a consumer perspective, but for the most part this is easily patched like this:

```diff
- $server = new Server($loop);
- $server->listen($port);
+ $server = new Server($port, $loop);
```

Also, empirical evidence suggests many places simply consume a `ServerInterface` in order to react to the `connection` event - this very common API is not affected.

This means that the `Server` class now starts listening immediately and a consumer can no longer defer the `listen()` call. I personally consider this a good thing (poka yoke) as I know many people have stumbled over this. Also, this API is now somewhat similar to Node's `net.createServer()` which also starts up a listening server (vs. the older explicit `server.listen()` call).

If you want to review this PR and don't want to mess with all these details, I'd suggest taking a look at the examples (and their minimal diff). The previous CLI accepted merely an optional port, the new CLI also accepts full listening addresses, for example like this:

```bash
$ php examples/02-chat.php 0.0.0.0:8080
```

Closes #19 
Supersedes / closes #6, refs https://github.com/reactphp/socket-client/pull/74, https://github.com/reactphp/react/issues/199 and #34


src/Server.php
```diff
@@ -83,16 +130,22 @@ public function handleConnection($socket)
...
 
         return (int) substr(strrchr($name, ':'), 1);
     }
 
     public function shutdown()
     {
-        $this->loop->removeStream($this->master);
-        fclose($this->master);
-        $this->removeAllListeners();
+        if (is_resource($this->master)) {
```
jsor: Minor nitpicking (and personal preference): I'd like to see an early return here.

```php
if (!is_resource($this->master)) {
    return;
}

$this->loop->removeStream($this->master);
fclose($this->master);
$this->removeAllListeners();
```

PR #66, opened 2017-01-27
Reject listening on hostnames and document valid URIs

### LINKED ISSUE #7: Server::listen() is blocking if given a hostname
> Migrated from https://github.com/reactphp/react/issues/181, originally reported 2013-04-13

Calling Socket\Server::listen($port, $host) with a hostname instead of an IPv4 address is a blocking operation.

Example:

``` php
$server = new Socket\Server($loop);
$server->listen(1234, 'example.com');
```

Giving a hostname will result in resolving it via DNS (blocking).

clue:
Listening on a hostname such as `example.com` or `localhost` is a blocking operation. This simple PR enforces the host to consists of an IP address.

Fixes / closes #7
Refs / builds on top of #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,13 +151,38 @@ is responsible for accepting plaintext TCP/IP connections.
 $server = new Server(8080, $loop);
 ```
 
-By default, the server will listen on the localhost address and will not be
-reachable from the outside.
-You can change the host the socket is listening on through a first parameter 
-provided to the constructor:
+As above, the `$uri` parameter can consist of only a port, in which case the
+server will default to listening on the localhost address `127.0.0.1` and thus
+it will not be reachable from outside of this system.
+
+In order to use a random port assignment, you can use the port `0`:
+
+```php
+$server = new Server(0, $loop);
+$port = $server->getPort();
+```
+
+In order to change the host the socket is listening on, you can provide an IP
+addres through the first parameter provided to the constructor, optionally
+preceded by the `tcp://` scheme:
 
 ```php
 $server = new Server('192.168.0.1:8080', $loop);
+```
+
+If you want to listen on an IPv6 address, you MUST enclose the host in square
+brackets:
+
+```php
+$server = new Server('[::1]:8080', $loop);
+```
+
+If the given URI is invalid, does not contain a port, any other scheme or if it
+contains a hostname, it will throw an `InvalidArgumentException`:
+
+```php
+// throws InvalidArgumentException due to missing port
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -151,15 +151,40 @@ is responsible for accepting plaintext TCP/IP connections.
...
+
+In order to use a random port assignment, you can use the port `0`:
+
+```php
+$server = new Server(0, $loop);
+$port = $server->getPort();
+```
+
+In order to change the host the socket is listening on, you can provide an IP
+addres through the first parameter provided to the constructor, optionally
```
jsor: missing s on *address*

src/Server.php
```diff
@@ -50,15 +50,40 @@ class Server extends EventEmitter implements ServerInterface
      * $server = new Server(8080, $loop);
      * ```
      *
-     * By default, the server will listen on the localhost address and will not be
-     * reachable from the outside.
-     * You can change the host the socket is listening on through the first parameter
-     * provided to the constructor.
+     * As above, the `$uri` parameter can consist of only a port, in which case the
+     * server will default to listening on the localhost address `127.0.0.1` and thus
+     * it will not be reachable from outside of this system.
```
jsor: Same as in https://github.com/reactphp/socket/pull/66/files#r98213995

src/Server.php
```diff
@@ -50,15 +50,40 @@ class Server extends EventEmitter implements ServerInterface
...
+     *
+     * In order to use a random port assignment, you can use the port `0`:
+     *
+     * ```php
+     * $server = new Server(0, $loop);
+     * $port = $server->getPort();
+     * ```
+     *
+     * In order to change the host the socket is listening on, you can provide an IP
+     * addres through the first parameter provided to the constructor, optionally
```
jsor: missing s on *address*

README.md
```diff
@@ -151,15 +151,40 @@ is responsible for accepting plaintext TCP/IP connections.
 $server = new Server(8080, $loop);
 ```
 
-By default, the server will listen on the localhost address and will not be
-reachable from the outside.
-You can change the host the socket is listening on through a first parameter 
-provided to the constructor:
+As above, the `$uri` parameter can consist of only a port, in which case the
+server will default to listening on the localhost address `127.0.0.1` and thus
+it will not be reachable from outside of this system.
```
jsor: superfluous *it* (not sure)? Maybe: *and thus will not be reachable*...

clue:
Updated to fix typos :shipit:

PR #68, opened 2017-02-05
Add getLocalAddress() method

clue:
This PR implements a simple `getLocalAddress()` method which complements the existing `getRemoteAddress()` method already present on the `ConnectionInterface`.

This can be useful if your `Server` instance is listening on multiple interfaces (e.g. using the address `0.0.0.0`). You can use this method to find out which interface actually accepted this connection (such as a public or local interface).

Adding this method is a new feature from a consumer perspective. Marking this as "BC break" in case any consumers of this package *implement* this interface (this should not affect most users, aka. *normal* usage).

Refs reactphp/react#199 and #65


clue:
Rebased now that #67 is in :shipit:

PR #76, opened 2017-02-21
Update README to use fully qualified class names

dandelionred:
Class namespace missed in an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Here is a server that closes the connection if you send it anything:
 $loop = React\EventLoop\Factory::create();
 
 $socket = new React\Socket\Server(8080, $loop);
-$socket->on('connection', function (ConnectionInterface $conn) {
+$socket->on('connection', function (React\Socket\ConnectionInterface $conn) {
     $conn->write("Hello " . $conn->getRemoteAddress() . "!\n");
     $conn->write("Welcome to this amazing server!\n");
     $conn->write("Here's a tip: don't say anything.\n");


clue:
Ping @reactphp/core, do we want to use fully qualified class names throughout all our READMEs?

jsor:
I'm not sure. I think the pattern is mostly that for classes from the package itself, we're *not* using fully qualified class names, only for classes from foreign packages.

In this particular case, there is `React\Socket\Server` used in the line above. So, for consistency one of the lines should be changed.

dandelionred:
I created the issue because the sample code didn't work copy-pasted. I assume examples must be self-contained so they are copy-paste runnable in any environment (i.e. they should not require any `use` statements in the code somewhere above).

kelunik:
How about just adding `use` statements instead of adding full qualified class names?

clue:
> I created the issue because the sample code didn't work copy-pasted.

It's my understanding that examples in the README serve as a way to get a basic overview how each component works within its ecosystem. These examples often omit certain things in order to highlight other aspects. The surrounding text includes hints as to what is missing and your IDE may guide you with adding missing imports.

Even if we were to use fully qualified class names as suggested by this PR throughout the README, this still wouldn't work when within a namespace. Using FQCN within production code is also a design flaw (arguably) and certainly not something we would want to endorse. Adding `use` statements everywhere would be an option, but this wouldn't make copy-pasting things any easier exactly.

Long story short, if you want to copy-paste working examples, please use the fully working examples from our `examples/` folder which can be run as-is. The examples provided in this folder have been improved somewhat since this ticket was originally filed, so I believe this to provide a better alternative nowadways.

I believe this has been answered, so I'm closing this for now. Please come back with more details if this problem persists and we can reopen this :+1: Thank you for your effort and sparking this discussion!

PR #78, opened 2017-03-01
Add support for unix:// transport

s2x:
This change adds support for unix domain socket transport for using unix:// for connections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,12 @@ brackets:
 $server = new Server('[::1]:8080', $loop);
 ```
 
+In order to use a unix domain socket (UDS), preceded socket path with `unix://`:
+
+```php
+$server = new Server('unix:///path/to/socket.sock', $loop);
+```
+
 If the given URI is invalid, does not contain a port, any other scheme or if it
 contains a hostname, it will throw an `InvalidArgumentException`:
 


clue:
While this may look good functionally, it unfortunately lacks documentation and tests. I've prepared a very much similar PR, do you feel like updating or should I file mine? :+1:

s2x:
I added some test, but hhvm fails. I will try to fix it tomorrow.

s2x:
Ok, i made ugly hack for hhvm because there is a bug in hhvm (see https://github.com/facebook/hhvm/issues/7733). 

Tomorrow I will add some info in readme about the unix sockets.

src/Server.php
```diff
@@ -206,7 +241,18 @@ public function close()
...
+
+        //get socket meta_data before closing
+        $metaData = stream_get_meta_data($this->master);
+        $socket_path = stream_socket_get_name($this->master, false);
         fclose($this->master);
+
+        //if unix socket we must delete socket file
+        if ($metaData['stream_type'] == 'unix_socket') {
+            //ugly hhvm hack see bug https://github.com/facebook/hhvm/issues/7733
+            unlink(rtrim($socket_path,":"));
```
clue: Shouldn't this workaround only be applied to affected versions?

src/Server.php
```diff
@@ -206,7 +241,18 @@ public function close()
         }
 
         $this->loop->removeStream($this->master);
+
+        //get socket meta_data before closing
+        $metaData = stream_get_meta_data($this->master);
+        $socket_path = stream_socket_get_name($this->master, false);
         fclose($this->master);
+
+        //if unix socket we must delete socket file
```
clue: I agree that this probably makes sense in most cases, but is this really a "must"?

src/Server.php
s2x: I will prepare new PR for this

src/Server.php
s2x: We need to delete socket file, because the socket file will remain after server close.

PR #80, opened 2017-03-09
add getMasterStream method

Tatikoma:
I'm writing forking TCP server with master and forked worker processes. I need the way to detach all socket listeners after process forked.
I see no other way to do it (except using Reflection), so would be good to add this functional.

Sample code when i forced to use Reflection: https://github.com/Tatikoma/react-mst/blob/b80be508093f09ae7e6d82a32274c7e4b5c948fa/src/tatikoma/react-mst/Server.php#L195

Or maybe i missed something then i 'll be happy if someone 'll show me other way :-)

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -181,6 +181,13 @@ final class Server extends EventEmitter implements ServerInterface
         });
     }
 
+    /**
+     * @return resource master stream
+     */
+    public function getMasterStream(){
+        return $this->master;
+    }
+    
     public function getAddress()
     {
         if (!is_resource($this->master)) {


clue:
Hi @Tatikoma, thanks for taking the time to file this PR! :+1: 

Unfortunately, the PR does not contain any documentation or tests, so it's a bit unclear what you're trying to achieve from the snippet you've posted.

May I ask you provide a simple gist of what you're trying to achieve? Thanks!

clue:
Closing this due to a lack of feedback. Please come back with more details if you think this is still relevant and we can reopen this :+1:

PR #85, opened 2017-03-27
Add link for previous version docs

Nek-:
You have a good velocity in development theses days. But some projects use a previous version and the doc looks wrong but it's not!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,11 @@
 
 [![Build Status](https://secure.travis-ci.org/reactphp/socket.png?branch=master)](http://travis-ci.org/reactphp/socket)
 
+For previous versions, please see related documentation:
+
+| [0.3.x](https://github.com/reactphp/socket/tree/0.3/README.md) | [0.4.x](https://github.com/reactphp/socket/tree/0.4/README.md) | [0.5.x](https://github.com/reactphp/socket/tree/master/README.md) |
+|---|---|---|
+
 Async, streaming plaintext TCP/IP and secure TLS socket server for React PHP
 
 The socket component provides a more usable interface for a socket-layer


WyriHaximus:
Thanks for filing this PR, you've made an excellent point there 👍 . Currently I'm not 100% sold on the location, personally I think it would be better to put it right above or right below the ToC. The first thing you read now is where to find older versions of the docs rather then what this package does.

Nek-:
Please wait this is badly formatted ! ^^

Nek-:
@WyriHaximus here it is with a better formatting. And the inspiration of formatting comes from guzzle : https://github.com/guzzle/guzzle#version-guidance

clue:
Thanks for filing this PR @Nek-, I agree with @WyriHaximus that you're making a very good point here :+1: Do you feel this is still relevant now that development has slowed down a bit and this component is now deprecated (#86)?

jsor:
I'd rather add something like that to reactphp.org, so i'm 👎 for adding this so prominently to the readme.

The following screenshot is from a local branch which i'm going to do a PR for reactphp/website when i find some time to finish it.

bildschirmfoto 2017-09-04 um 21 22 09

Nek-:
Looks like a good idea. Only if you redirect the user directly to reactphp.org from the README. Which is not the case as you actually uses the README to generate reactphp.org.

This makes me sceptic and may still create a lot of frustration. (that was mine before I created this PR haha)

... Especially since the development is so heavy. :D cc  @clue

clue:
@jsor's suggestion from above has been implemented via https://github.com/reactphp/website/pull/20 :shipit: 

I believe this has been answered, so I'm closing this for now. Please come back with more details if this problem persists and we can reopen this :+1:

PR #86, opened 2017-04-02
Add LimitingServer to limit and keep track of open connections

clue:
Add `LimitingServer` to limit number of open connections. This can be useful to limit how many resources are spent on handling connections vs. handling the requests that arrive on each connection (HTTP).

Builds on top of #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ and [`Stream`](https://github.com/reactphp/stream) components.
     * [close()](#close)
   * [Server](#server)
   * [SecureServer](#secureserver)
-  * [AccountingServer](#accountingserver)
+  * [LimitingServer](#limitingserver)
     * [getConnections()](#getconnections)
   * [ConnectionInterface](#connectioninterface)
     * [getRemoteAddress()](#getremoteaddress)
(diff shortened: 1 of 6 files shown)


clue:
Updated to work around a legacy PHP 5.3 issue due to old react/event-loop :shipit:

jsor:
Out of curiosity: Is `Accounting` a common term for this? Because when i've read the notification of this PR, i've been very confused by the name ;)

clue:
Afaict "accounting" is not a common name for this feature and I'm open for suggestions :-) Afaict, this is not exactly a common feature in the first place, because this is often implemented as part of the underlying "server" instead of via decoration. Given our current design, I would vote to keep this via decoration (e.g. this allows us to limit the number of TLS connections), but I'm not opposed to renaming this feature :+1: 

FWIW, the name "accounting" is related to "accountant" or "book keeping", which I think is good fit for this feature, because it does more than just "limiting".

jsor:
> FWIW, the name "accounting" is related to "accountant" or "book keeping", which I think is good fit for this feature, because it does more than just "limiting".

Yeah, i figured after seeing the code, but looked strange just from the name ;)

What about `ConnectionControllingServer`?

clue:
Alright, renamed AccountingServer to LimitingServer, as I think confusion with "managing accounts" is a valid concern :+1: :shipit:

PR #93, opened 2017-04-17
Check for empty socket address

### LINKED ISSUE #92: Handle stream_socket_get_name return "\0"
https://github.com/reactphp/socket/blob/master/src/Connection.php#L44 should not only check for `false`but also for `"\0"` according to https://bugs.php.net/bug.php?id=74458.

Discovered through https://github.com/reactphp/http/pull/160#issuecomment-294438538. Since ratchet still uses 0.5 it would be great if this could also be back ported to 0.5...0.7.

andig:
Fix #92

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -41,7 +41,8 @@ class Connection extends Stream implements ConnectionInterface
 
     private function parseAddress($address)
     {
-        if ($address === false) {
+        // work around https://bugs.php.net/bug.php?id=74458 by checking if  stream_socket_get_name has returned null string instead of address
+        if ($address === false || $address === "\0") {
             return null;
         }
 


clue:
@andig Thanks for taking the time to file this PR! :+1: 

Unfortunately, the PR does not contain any documentation or tests, so it's a bit unclear what you're trying to achieve from the snippet you've posted.

May I ask you provide a simple gist of what you're trying to achieve, i.e. how can this be reproduced and how can we be sure this issue is actually resolved and we do not introduce any regressions in the future? Thanks!

andig:
> May I ask you provide a simple gist of what you're trying to achieve, i.e. how can this be reproduced and how can we be sure this issue is actually resolved and we do not introduce any regressions in the future? Thanks!

This happens from inside our application. Unfortunately I cannot produce a simple gist. Looking at lxr I believe returning the NULL string is what PHP does- under which condition whatsoever.

I've added comment to the code what the check is supposed to do. There was no parseAddress test before and I have unfortunately no clue how to write one- sorry.

andig:
Thanks for fixing. Now- for sake of documentation- I understand that the observed behaviour happens with UNIX domain sockets.

PR #100, opened 2017-05-08
Fix unix:// addresses for Unix domain socket (UDS) paths

### LINKED ISSUE #92: Handle stream_socket_get_name return "\0"

clue:
Resolves / closes #92
Supersedes / closes #93
Builds on top of #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,8 @@ If the remote address can not be determined or is unknown at this time (such as
 after the connection has been closed), it MAY return a `NULL` value instead.
 
 Otherwise, it will return the full address (URI) as a string value, such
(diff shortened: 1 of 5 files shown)


clue:
Updated to fix issues with HHVM < 3.19 (https://github.com/facebook/hhvm/issues/7733) :shipit:

PR #109, opened 2017-06-12
Fix HHVM build for now again and ignore future HHVM build errors

clue:
The HHVM build reports an error once again. Let's fix this once again by updating the base distro from precise to trusty and ignore any future HHVM build errors.

See https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-300553336
Originally from https://github.com/clue/php-http-proxy-react/pull/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,14 @@ php:
   - 5.5
   - 5.6
   - 7
-  - hhvm
+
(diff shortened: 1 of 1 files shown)


clue:
For the reference: The build output reveals that HHVM now fails to establish any TLS sessions due to a timeout (it even fails to send an HTTPS request to Google) and I have no plans to spend any time on this right now.

PR #125, opened 2017-10-08
Fix forward compatibility with upcoming EventLoop releases in tests

clue:
This fixes a small oversight from #104: The test suite still used the deprecated `tick()` method which will be removed in the upcoming EventLoop release.

Builds on top of #104

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "react/promise-timer": "~1.0"
     },
     "require-dev": {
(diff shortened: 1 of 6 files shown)


andig:
@clue you forgot https://github.com/reactphp/socket/blob/master/tests/TcpServerTest.php#L69 ⭕️

clue:
@andig Duh! Thanks for spotting, see #127 :shipit:

PR #128, opened 2017-10-12
Fix secure server test

the-eater:


src/SecureServer.php
```diff
@@ -141,7 +141,14 @@ public function __construct(ServerInterface $tcp, LoopInterface $loop, array $co
...
     {
-        return str_replace('tcp://' , 'tls://', $this->tcp->getAddress());
+        $tcpAddress = $this->tcp->getAddress();
+        $schemePos = strpos($tcpAddress, '://');
+
+        if ($schemePos === false) {
+            return 'tls://' . $tcpAddress;
+        }
+
+        return 'tls' . substr($tcpAddress, $schemePos);
```
clue: Note that this PR is fixing an issue that doesn't exist :-) The `TcpServer` will *always* prepend a `tcp://` scheme as documented in the README. I understand that the tests is misleading (and probably a small oversight from #98), maybe simply adjust this there?

kelunik: Indeed, that's an issue with the mock.

Issue #126, opened 2017-10-10
Support half-open connections (FIN)


clue:
This is particularly useful for some TCP/IP-based protocols where the client sends a FIN message to the server indicating that it will no longer write any data but is still willing to accept incoming data.

Opening this ticket as a reminder and to spark the discussion - the implementation likely depends on  https://github.com/reactphp/stream/issues/27

pronskiy:
Hi guys!
I started to work on this. While still not sure if I can finish, I think I got some results by changing `React\Stream\WritableResourceStream::close()` and replacing lines:
```
 if (is_resource($this->stream)) {
    fclose($this->stream);
}
```
with
```
@stream_socket_shutdown($this->stream, 1);
stream_set_blocking($this->stream, false);
```
i.e. keeping it readable in `DuplexResourceStream`.

Do you think this is a reasonable direction?

clue:
@pronskiy Thank you for looking into this! Sounds like this makes perfect sense on a lower level. This will likely cause a number of tests to fail, but this can probably shouldn't be too much work either. But I'm really curious how this could be exposed API-wise. Have you looked into this yet?

pronskiy:
@clue I plan to submit some draft for your consideration so that we have a basis for discussion.

pronskiy:
@clue here are couple PRs with an early draft. Basically, I add `$options` array as a parameter to various classes' costructor. It's kinda bad because the same array is passed from one class to another. I'm thinking of how to improve it. What do you think of such direction?

clue:
I still think that this feature makes perfect sense :+1: However, there are currently no immediate plans to build this from my end (no demand at the moment and more important outstanding issues currently), but I suppose we would be happy to accept PRs in case anybody feels there's a need for this :+1:

PR #131, opened 2017-11-05
[Draft] Support half-open connections (FIN)

pronskiy:
#126

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -40,9 +40,12 @@ class Connection extends EventEmitter implements ConnectionInterface
     public $stream;
 
     private $input;
+    private $options;
 
-    public function __construct($resource, LoopInterface $loop)
+    public function __construct($resource, LoopInterface $loop, array $options = array())
     {
+        $this->options = $options;
+
         // PHP < 5.6.8 suffers from a buffer indicator bug on secure TLS connections
         // as a work-around we always read the complete buffer until its end.
         // The buffer size is limited due to TCP/IP buffers anyway, so this
@@ -66,7 +69,8 @@ class Connection extends EventEmitter implements ConnectionInterface
             $resource,
             $loop,
             $clearCompleteBuffer ? -1 : null,
-            new WritableResourceStream($resource, $loop, null, $limitWriteChunks ? 8192 : null)
+            new WritableResourceStream($resource, $loop, null, $limitWriteChunks ? 8192 : null, $this->options),
+            $this->options
         );
 
         $this->stream = $resource;
--- a/src/TcpConnector.php
+++ b/src/TcpConnector.php
@@ -11,10 +11,11 @@ final class TcpConnector implements ConnectorInterface
     private $loop;
     private $context;
 
-    public function __construct(LoopInterface $loop, array $context = array())
+    public function __construct(LoopInterface $loop, array $context = array(), array $options = array())
     {
         $this->loop = $loop;
         $this->context = $context;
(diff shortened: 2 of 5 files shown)


tests/AllowHalfOpenTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+use React\EventLoop\Factory;
+use React\EventLoop\LoopInterface;
+use React\Socket\ConnectionInterface;
+use React\Socket\Connector;
+use React\Socket\TcpServer;
+
+function _log ($message) {
+    echo $message . "\n";
+};
```
kelunik: Why not just use `echo` directly?

pronskiy: You are right, it's because _log function was more complex before. Anyway, I plan to remove it when the API is discussed and the PR is ready. Please let me know what do you think of approach in general #126?

src/TcpConnector.php
```diff
@@ -11,10 +11,11 @@
     private $loop;
     private $context;
 
-    public function __construct(LoopInterface $loop, array $context = array())
+    public function __construct(LoopInterface $loop, array $context = array(), array $options = array())
```
clue: Not a big fan of these "options" tbh, how about transporting this as part of "context"? Note that this would also need to be covered in our README for `Connector`, `TcpConnector`, `Server` and `TcpServer` at least.

src/Connection.php
```diff
@@ -40,9 +40,12 @@ class Connection extends EventEmitter implements ConnectionInterface
     public $stream;
 
     private $input;
+    private $options;
 
-    public function __construct($resource, LoopInterface $loop)
+    public function __construct($resource, LoopInterface $loop, array $options = array())
     {
+        $this->options = $options;
```
clue: See below, I think we *may* remove this parameter if we transport this as part of the context of the resource.

src/TcpServer.php
```diff
@@ -123,6 +124,7 @@
     public function __construct($uri, LoopInterface $loop, array $context = array())
     {
         $this->loop = $loop;
+        $this->context = $context;
```
clue: The context for each connection is inherited from the server socket context, is this assignment really needed?

tests/AllowHalfOpenTest.php
```diff
@@ -0,0 +1,143 @@
...
+    public function setUp()
+    {
+        $this->loop = $this->createLoop();
+        $this->socket = new TcpServer(0, $this->loop, $this->options);
+    }
+
+    /**
+     * @covers React\Socket\TcpServer::handleConnection
+     */
+    public function testAll()
```
clue: This is apparently supposed to test the server sending a `FIN` to the client which then sends a `FIN` in return?

I think we should add another test for a client sending a `FIN` to the server which then sends a `FIN` in return.

Also, I don't see this covered in the code, have you verified the `FIN` flags are actually sent?

pronskiy:
@clue thanks a lot for the thorough review! I will need some time to address the issues.

clue:
Again, thank you for digging into this and filing these PRs! :+1:

I still think that getting this feature in makes perfect sense and I'm looking forward to a future update here!

Given that this is currently a WIP and hasn't seen any updates in a while and I don't see it's likely this will get traction any time soon, I will have to close this for now. Please don't see this as discouraging, we would still love to get this feature in and very much appreciate your effort! You can simply complete this feature and update this PR and we're happy to reopen this :+1:

Issue #132, opened 2017-11-09
Need to upgrade phpunit for php 7.2


andig:
Starting to see these notices:

	❯ vendor/bin/phpunit tests/ServerTest.php
	PHP Deprecated:  The each() function is deprecated. This message will be suppressed on further calls in /Users/andig/Documents/htdocs/react-socket/vendor/phpunit/phpunit/src/Util/Getopt.php on line 38

	Deprecated: The each() function is deprecated. This message will be suppressed on further calls in /Users/andig/Documents/htdocs/react-socket/vendor/phpunit/phpunit/src/Util/Getopt.php on line 38
	PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

clue:
Thanks for reporting! Can you show the output of `composer info`? Also, have you tried running `composer update`?

https://github.com/sebastianbergmann/phpunit/issues/2472 suggests this should be omitted with PHPUnit 5.7.11

andig:
It seems though that even 5.7.11 only suppresses the notice but does not fix the root cause. This will likely break the tests again once PHP >7.2 has been released and `each` removed. At this point there will not be a phpUnit version that supports PHP 5.6 and will have the required fix?

kelunik:
It will take some more years for PHP 8.0 to be released. At that point, support for PHP 5.x can probably be dropped. See https://wiki.php.net/rfc/deprecations_php_7_2#each

clue:
@andig can you confirm that using PHPUnit 5.7.11 fixes this issue? Also, do you see the same issue with the latest PHPUnit ^4.8? Can we even give a consistent `require-dev.phpunit = "^5.7.11 || ^4.8"` that works across all supported PHP versions?

andig:
No. Stuck in dependency hell (master branch) and no time digging deeper right now:


    ❯ composer require phpunit/phpunit:5.7.11
    ./composer.json has been updated
    Loading composer repositories with package information
    Updating dependencies (including require-dev)
    Your requirements could not be resolved to an installable set of packages.

      Problem 1
        - react/socket v0.7.2 requires react/stream ^0.6 || ^0.5 || ^0.4.5 -> satisfiable by react/stream[0.4.x-dev, v0.4.5, v0.4.6, v0.5.0, v0.6.0] but these conflict with your requirements or minimum-stability.
[...]

PR #133, opened 2017-11-12
Forward compatibility with PHPUnit 6

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCase. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06) and [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -98,7 +98,6 @@ class IntegrationTest extends TestCase
         $this->assertNotRegExp('#^HTTP/1\.0#', $response);
     }
 
-    /** @test */
     public function testConnectingFailsIfDnsUsesInvalidResolver()
     {
         $loop = Factory::create();
@@ -114,7 +113,6 @@ class IntegrationTest extends TestCase
         Block\await($connector->connect('google.com:80'), $loop, self::TIMEOUT);
     }
 
-    /** @test */
     public function testConnectingFailsIfTimeoutIsTooSmall()
     {
         if (!function_exists('stream_socket_enable_crypto')) {
@@ -131,7 +129,6 @@ class IntegrationTest extends TestCase
         Block\await($connector->connect('google.com:80'), $loop, self::TIMEOUT);
     }
 
-    /** @test */
     public function testSelfSignedRejectsIfVerificationIsEnabled()
     {
         if (!function_exists('stream_socket_enable_crypto')) {
@@ -150,7 +147,6 @@ class IntegrationTest extends TestCase
         Block\await($connector->connect('tls://self-signed.badssl.com:443'), $loop, self::TIMEOUT);
     }
 
-    /** @test */
     public function testSelfSignedResolvesIfVerificationIsDisabled()
     {
         if (!function_exists('stream_socket_enable_crypto')) {
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -163,6 +163,9 @@ class IntegrationTest extends TestCase
 
         $conn = Block\await($connector->connect('tls://self-signed.badssl.com:443'), $loop, self::TIMEOUT);
         $conn->close();
+
+        // if we reach this, then everything is good
+        $this->assertNull(null);
     }
 
     public function testCancelPendingConnection()
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "clue/block-react": "^1.2",
-        "phpunit/phpunit": "^5.0 || ^4.8"
+        "phpunit/phpunit": "^5.7 || ^4.8.35"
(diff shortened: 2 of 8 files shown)


andig:
According to https://github.com/reactphp/socket/issues/132 it should be 5.7.11 as min version.

gabriel-caruso:
@andig What do you suggest: `5.7`, `5.7.11` or the latest stable [`5.7.23`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#5723---2017-10-15)?

clue:
@andig I think that this covers a different issue, so I would vote to keep this separated for now? :+1:

gabriel-caruso:
@clue Added :smile:

tests/IntegrationTest.php
```diff
@@ -98,7 +98,10 @@ public function gettingPlaintextStuffFromEncryptedGoogleShouldNotWork()
         $this->assertNotRegExp('#^HTTP/1\.0#', $response);
     }
 
-    /** @test */
+    /**
+     * @test
+     * @expectedException RuntimeException
```
jsor: Be aware that this change has implications for the test. Now, all `RuntimeException`s are expected (eg. from the setup code) while before, only exceptions thrown from `connect()` were expected (I'm aware this change has been made for being compatible with PHPUnit 6).

Maybe test with `@expectedExceptionMessage` explicitely to avoid covering other exceptions?

gabriel-caruso: Going to add `@expectedExceptionMessage` in all changes to double check!

kelunik: You could also simply go with something like https://github.com/sebastianbergmann/phpunit/issues/2074#issuecomment-186322116 instead.

clue: @kelunik suggestion may have the benefit of resulting in a smaller changeset (with less potential for subtle issues as below), but this would still require some additional logic to skip this on newer PHPUnit versions.

Either way, I'm fine with either solution :+1:

tests/IntegrationTest.php
```diff
@@ -110,11 +113,18 @@ public function testConnectingFailsIfDnsUsesInvalidResolver()
             'dns' => $dns
         ));
 
-        $this->setExpectedException('RuntimeException');
-        Block\await($connector->connect('google.com:80'), $loop, self::TIMEOUT);
+        $result = Block\await($connector->connect('google.com:80'), $loop, self::TIMEOUT);
+
+        $this->expectExceptionMessage(
+          'DNS query for google.com failed: Unable to connect to DNS server: php_network_getaddresses: getaddrinfo failed: ' . $result
+        );
```
clue: This will not be executed as the above method already throws. This is also the reason why it *may* make sense to keep the original test structure (see above discussion), but I'll leave this up to you :+1: 

(see also below occurrences)

clue: Ping @gabriel-caruso, what's the status here?

gabriel-caruso: @clue Sorry, I didn't see that. I gonna move `expect` methods to the beginning of the methods.

tests/IntegrationTest.php
gabriel-caruso: Guys, what should I do? `expectExceptionMessage` isn't available anymore on PHPUnit 6, and this assertion in variable :cry:

clue: How about this pseudo code? Maybe even move this to a `setExpectException()` helper method for backwards compatibility as @kelunik suggested?
```
if (phpunit >= 6) {
    $this->expectException('RuntimeException');
    $this->expectExceptionMessage("hello");
} else {
    $this->setExpectException('RuntimeException', 'hello');
}
```

gabriel-caruso:
Ok, we are facing some issues with this update. I have some solutions, and I want your opinion:

gabriel-caruso:
Ok, we are facing some issues with this update. I have some solutions, and I want your opinion:

- Just update to PHPUnit 5, because PHPUnit has **big** breaking changes;
- Drop PHPUnit 4 support, but require minimum `PHP 5.6`;

clue:
@gabriel-caruso Thank you for keeping up with this! I've updated this PR to add a BC layer as per the above discussion so that all PHPUnit versions are now supported :tada:

Issue #150, opened 2018-02-13
Allow getRemoteAddress to return the PID of a connecting Unix connection


gdejong:
Currently when dealing with a `UnixServer`, its connections do not return a value for `getRemoteAddress`, the returned value is actually `NULL`.
I`d like it to return the PID of the connecting client. Is that something you would integrate in this library?

    public function getRemoteAddress()
    {
        if($this->unix){
            $socket = socket_import_stream($this->stream);

            return socket_get_option($socket, SOL_SOCKET, 17 /*SO_PEERCRED*/);
        }
        return $this->parseAddress(@stream_socket_get_name($this->stream, true));
    }

See http://php.net/manual/en/function.socket-get-option.php#101380

clue:
Absolutely! I like the idea of returning the PID and can see some very valid use cases (local authentication etc.).

If we can integrate this into this library without introducing a BC break and/or a hard dependency on `ext-sockets`, then I see no reason why we wouldn't want to get this in :+1: PRs with added tests and documentation would be much appreciated :shipit:

gdejong:
I'd like to ask a few questions before I try to create PR for this :)
- The `getRemoteAddress` method from the `ConnectionInterface` interface should return the remote address (URI) or null if unknown. In this case only the PID is known, which is an `integer`. What should we return? There isn't really a URI scheme for a PID as far as I know.
- Another solution would be to add another method to a `Connection`, like `getUnixPid` (name definitely to be decided). For all non-unix connections this would return `null`, and for unix connections it would return the pid as an integer.

What are your thoughts on this?

clue:
Very good points!

Interestingly `SO_PEERCRED` socket option should return a `ucred` structure with `pid`, `uid` and `gid`. Right now, PHP does not really support this constant and simply seems to return the first element only.

A single access may work for now, but I can clearly see some value in the remaining values once PHP actually implements this structure.

As an alternative, what do you think about exposing this as part of the query string?

```
$client->getRemoteAddress();
unix:///tmp/server.sock?pid=1&uid=2&gid=3
```

kelunik:
Depending on the parser you can also expose the PID as port number.

gdejong:
I have created the most basic Unix server/client to test this feature with. Leaving it here for future reference.
### server.php
```
<?php

require __DIR__ . "/vendor/autoload.php";

use React\EventLoop\Factory;
use React\Socket\ConnectionInterface;
use React\Socket\UnixServer;

[...]

PR #153, opened 2018-03-31
Created a UnixConnection to represent UDS connections which has the ability to retrieve the local/remote PID

gdejong:
Connections through a Unix Domain Socket will now be represented by a `UnixConnection` (which extends a regular `Connection`). This `UnixConnection` contains the custom logic for parsing parsing the address and has the new ability to retrieve the local and remote PID.

From the server side a client has no "remote address". The remote PID can for example be used to distinguish multiple clients.

See #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ handle multiple concurrent connections without blocking.
   * [ConnectionInterface](#connectioninterface)
     * [getRemoteAddress()](#getremoteaddress)
     * [getLocalAddress()](#getlocaladdress)
+  * [UnixConnection](#connectioninterface)
+    * [getRemotePid()](#getremotepid)
+    * [getLocalPid()](#getlocalpid)
 * [Server usage](#server-usage)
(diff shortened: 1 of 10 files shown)


dkrieger:
What's the status on this? Was BC break label added as precaution, or is compatibility truly broken? If the issue is no longer setting the 'unix' public property to true, there's no reason that can't still be done in this PR. Let me know if there's anything I can do to help get this feature merged.

gdejong:
Please let me know if there is something I can do to help with this PR :)

clue:
As much as I'd love to get this feature in, I'm having to close this for now as it hasn't received any input in a while and it's unlikely this will get traction any time soon.

The feature request is still open in #150 and we'll look into this again in the not too far future :+1: If you feel this was closed prematurely or want to pick this up again, please let us know and we can reopen this.

Thank you for your effort @gdejong, keep it coming!

PR #161, opened 2018-05-05
Improve memory consumption for failed connection attempts

clue:
While debugging some very odd memory issues in a live application, I noticed that this component shows some unexpected memory consumption and memory would not immediately be freed as expected. Let's not call this a "memory leak", because memory was eventually freed, but this clearly caused some unexpected and significant memory growth.

Parts of this issue have been resolved with #159 already which raised the discussion to address these issues in the upstream react/promise component. A number of patches have been filed via https://github.com/reactphp/promise/pull/113, https://github.com/reactphp/promise/pull/115, https://github.com/reactphp/promise/pull/116, https://github.com/reactphp/promise/pull/117, https://github.com/reactphp/promise/pull/118 and https://github.com/reactphp/promise-timer/pull/33.

I'm marking this PR as WIP because the referenced changes are yet to be released as part of react/promise v2.6.0. Once this release is out, I'll update the version reference and this should be ready to be shipped.

This PR actually includes a number of tests that show how garbage memory references are no longer an issue in any supported PHP version and how failed connection attempts no longer cause any such references on their own (unlike the previous changes in #159 this means that we can now test this and this requires no effort on the consumer side anymore).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,8 +9,8 @@
         "react/dns": "^0.4.13",
         "react/event-loop": "^1.0 || ^0.5 || ^0.4 || ^0.3.5",
         "react/stream": "^1.0 || ^0.7.1",
-        "react/promise": "^2.1 || ^1.2",
-        "react/promise-timer": "~1.0"
+        "react/promise": "^2.6.0 || ^1.2.1",
+        "react/promise-timer": "^1.4.0"
     },
     "require-dev": {
         "clue/block-react": "^1.2",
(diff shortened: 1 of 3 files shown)


clue:
For the reference, the one failing test case covers the problem addressed in https://github.com/reactphp/promise-timer/pull/33 and reactphp/promise#117 (and also indirectly covered in reactphp/promise#118, but for PHP 7+ only), so that this should be fixed once these PRs are in. I'll update this PR accordingly then. :shipit:

clue:
Updated and removed WIP marker now that https://github.com/reactphp/promise/releases/tag/v2.6.0 and https://github.com/reactphp/promise-timer/releases/tag/v1.4.0 have been released, this is now ready :shipit:

PR #162, opened 2018-05-05
Fix Travis config to test against legacy PHP 5.3 again

clue:
Just noticed that legacy PHP 5.3 does not show up in the Travis build log any more while working on #161. I'm not suggesting this is a particularly important version, but then again fixing this was somewhat trivial. Apparently, this was caused by a duplicate YAML definition in #130 (yey).


clue:
I've just updated this to fix the test case failing on PHP 5.3 only introduced via #146 :shipit:

PR #169, opened 2018-07-19
Improve TLS error messages (Connection lost during TLS handshake)

clue:
PHP does not seem to report any warning message when the connection is lost (closed by the remote side) during a TLS handshake (both client and server side). As a result of this, the previous message appeared to be cut off. This PR ensures we indicate this error with proper errno/errstr.

Refs #168


clue:
Rebased on current master and updated to skip testing platform specific TLS error message strings :shipit:

PR #181, opened 2018-11-11
Use dedicated assertInstanceOf assertion

carusogabriel:
Using this dedicated assertion, we can have better error messages:
```diff
-$this->assertTrue($error instanceof \RuntimeException); // Failed asserting that false is true.
+$this->assertInstanceOf(\RuntimeException::class, $error); // Failed asserting that an object is an instance of class "RuntimeException".
```

### DIFF
--- a/tests/FunctionalSecureServerTest.php
+++ b/tests/FunctionalSecureServerTest.php
@@ -421,7 +421,7 @@ class FunctionalSecureServerTest extends TestCase
         $error = Block\await($errorEvent, $loop, self::TIMEOUT);
 
         // Connection from tcp://127.0.0.1:39528 failed during TLS handshake: Connection lost during TLS handshak
-        $this->assertTrue($error instanceof \RuntimeException);
+        $this->assertInstanceOf('RuntimeException', $error);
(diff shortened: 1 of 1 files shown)


tests/FunctionalSecureServerTest.php
```diff
@@ -421,7 +421,7 @@ public function testEmitsErrorIfConnectionIsClosedBeforeHandshake()
         $error = Block\await($errorEvent, $loop, self::TIMEOUT);
 
         // Connection from tcp://127.0.0.1:39528 failed during TLS handshake: Connection lost during TLS handshak
-        $this->assertTrue($error instanceof \RuntimeException);
+        $this->assertInstanceOf(\RuntimeException::class, $error);
```
jsor: `::class` PHP >= 5.5 only

kelunik:
At least PHPStorm also understands assertInstanceOf for static analysis IIRC.

afk11:
In general, IDE capabilities (which vary greatly from one to the other) shouldn't affect design decisions.. Adding something like this would make the IDE-static-analysis business a non-blocker IMO:

/** @var \RuntimeException $error */

Issue #189, opened 2019-01-22
TLS Server fails without explanation


siolfyr:
Using react/socket version 1.2.0, using php --version 7.0.32-0ubuntu0.16.04.1

Started a server as in the README via:
```
$server = new React\Socket\Server('tls://0.0.0.0:25', $loop, array(
    'tls' => array(
        'local_cert' => 'secret.pem'
    )
));
```

And created the self-signed cert using the method in the 99 self-sign example file, which should in theory be valid. Also tried with a LetsEncrypt certificate, and got the same result.

The result is a server which triggers this error when hitting it with `openssl s_client -connect site.com:25`:
error: Connection from tcp://etc:etc failed during TLS handshake: Unable to complete TLS handshake: Unable to set private key file `/blah/blah/secret.pem'

The file exists and the error isn't the same as an error for a nonexistent file.

The server worked great before attempting to enable TLS.

What could be the issue? Any tips on debugging the issue with the certificate?

clue:
That's strange, this shouldn't have happened.

Here's an excerpt from the [README](https://github.com/reactphp/socket#secureserver) on how to diagnose these issues:

> > Note that the certificate file will not be loaded on instantiation but when an incoming connection initializes its TLS context. This implies that any invalid certificate file paths or contents will only cause an error event at a later time.
>
> Whenever a client fails to perform a successful TLS handshake, it will emit an error event and then close the underlying TCP/IP connection:
>
> $server->on('error', function (Exception $e) {
    echo 'Error' . $e->getMessage() . PHP_EOL;
});

Other than that, how was this pem file created? Have you tried verifying this in another program that can load pem files? Have you verified the same error happens with the demo certificate that ships in the examples/ folder?

siolfyr:
Hmm, that's odd. I tried the localhost.pem from the examples folder just now and it worked fine.

The cert was generated using the example in the examples folder, like this:
`php 99-generate-self-signed.php site.com swordfish > secret.pem`

I also tried using a LetsEncrypt cert, which works for my website fine. I also tried using a cert created via this method (option 1):
https://workaround.org/ispmail/jessie/create-certificate

My openssl version is: OpenSSL 1.0.2g  1 Mar 2016

I'll try using the cert I made in Nginx and see if it has a problem with it.

clue:
> The cert was generated using the example in the examples folder, like this:
> `php 99-generate-self-signed.php site.com swordfish > secret.pem`

[...]

PR #190, opened 2019-01-22
Example 99 improvements

jsor:
Minor fixes and improvements for `examples/99-generate-self-signed.php` discovered in #189


jsor:
@clue Yes, i noticed that as well. But i didn't find any direct references to this script. It's still possible to generate a cert with the password `swordfish`.

clue:
@jsor Good, no objections then :+1:

PR #211, opened 2019-09-04
EOF check on invalid resource

CharlotteDunois:
During the debugging of TLS support for a MySQL driver, I've encountered a SSL error which made the stream resource close and thus emit an error when we checked whether there's an EOF.

Error on `feof` was `supplied resource is not a valid stream resource` due to `stream_socket_enable_crypto` already failing with the error `An existing connection was forcibly closed by the remote host`.

This PR therefore only checks EOF if the resource is still valid.

### DIFF
--- a/src/StreamEncryption.php
+++ b/src/StreamEncryption.php
@@ -127,7 +127,7 @@ class StreamEncryption
             $d = $deferred;
             $deferred = null;
 
-            if (\feof($socket) || $error === null) {
+            if ($error === null || (\is_resource($socket) && \feof($socket))) {
                 // EOF or failed without error => connection closed during handshake
                 $d->reject(new \UnexpectedValueException(
                     'Connection lost during TLS handshake',


CharlotteDunois:
@clue I've been thinking about this aswell, but I'm not sure how this can be reliably tested. Quickly testing whether closing the socket from the server is enough was negative.

My other idea would be starting a new process for the server and then killing the server process after getting a successful connect while trying to setup TLS, to get a TCP reset. However I've yet to get to test this and see if that works.

If you have an idea how this can be tested, I'd like to add that as test.

clue:
@CharlotteDunois Perhaps take a look at #169 which originally introduced this `feof()` check. It's my understanding that if the resource is closed already, the `stream_socket_crypto_enable()` call should also report a warning, so perhaps just switching the statements as in your original version would be preferred? In the test suite, we can just call the internal `toggleCrypto()` method with some invalid arguments, this should already trigger the faulty behavior.

PR #229, opened 2020-05-09
Skip legacy TLS 1.0 / TLS 1.1 tests if disabled by system

clue:
Skip legacy TLS 1.0 / TLS 1.1 tests if disabled by system. For example, this affects a current Ubuntu 20.04 installation which appears to prohibit legacy TLS connections.

The same bug can be reproduced in curl with `$ curl -s --tlsv1 -v https://tls-v1-0.badssl.com:1010/` at the time of writing this. This seems to be addressed in curl with https://github.com/curl/curl/issues/4097 and https://github.com/curl/curl/pull/4304. PHP appears to apply a similar logic, but it still fails at the moment (https://github.com/php/php-src/blob/8e9bc90004f09a0d67fdac393a8f3fca904be397/ext/openssl/xp_ssl.c#L1760 vs https://github.com/ruby/openssl/blob/4b43ffc1292eeb70ff886847836e21ad96ed8796/ext/openssl/ossl_ssl.c#L162-L195).

For PHP, this can be reproduced (prior to applying this patch) by running the test suite in a default Ubuntu installation (PHP 7.4.3):

```bash
$ docker run -it --rm -v `pwd`:/data --workdir=/data ubuntu:20.04 bash
# apt update && apt install -y php-cli php-xml php-mbstring ca-certificates
# vendor/bin/phpunit
```

### DIFF
--- a/tests/FunctionalSecureServerTest.php
+++ b/tests/FunctionalSecureServerTest.php
@@ -148,6 +148,44 @@ class FunctionalSecureServerTest extends TestCase
         $this->assertEquals('TLSv1.2', $meta['crypto']['protocol']);
     }
 
+    public function testClientUsesTls10WhenCryptoMethodIsExplicitlyConfiguredByClient()
+    {
+        if (PHP_VERSION_ID < 70000) {
+            $this->markTestSkipped('Test requires PHP 7+ for crypto meta data');
+        }
+
+        $loop = Factory::create();
+
+        $server = new TcpServer(0, $loop);
+        $server = new SecureServer($server, $loop, array(
+            'local_cert' => __DIR__ . '/../examples/localhost.pem'
+        ));
+
+        $connector = new SecureConnector(new TcpConnector($loop), $loop, array(
+            'verify_peer' => false,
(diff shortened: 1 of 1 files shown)


tests/FunctionalSecureServerTest.php
```diff
@@ -148,6 +148,44 @@ public function testClientUsesTls12WhenCryptoMethodIsExplicitlyConfiguredByServe
...
+
+            throw $e;
+        }
+
+        $this->assertInstanceOf('React\Socket\Connection', $client);
+        $this->assertTrue(isset($client->stream));
+
+        $meta = stream_get_meta_data($client->stream);
+        $this->assertTrue(isset($meta['crypto']['protocol']));
+        $this->assertEquals('TLSv1.1', $meta['crypto']['protocol']);
```
jsor, line 186: The test suggests this should be TLS v1.0 🤔

clue: Duh, that's what you get when skipping some of the tests on your platform. Fixed now :+1:

PR #242, opened 2020-07-31
Change default socket backlog size to 511

clue:
The `backlog` parameter used to default to just 32 as of PHP 5.3.3 and
used to be hard coded to 4 in earlier PHP versions.

The new default of 511 was evaluated to be a reasonable value for
event-driven applications that can cope with many concurrent connections
on a variety of platforms. In particular, this solves common ECONNRESET
(connection reset by peer) errors when running benchmarks with higher
concurrency. For more specific requirements, the `backlog` parameter can
be given explicitly to override this default.

See https://github.com/php/php-src/commit/5b277c4da0d9fb8bfb59186b61a884aaafffb861
Refs https://github.com/reactphp/http/issues/354 and others

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,7 @@ $server = new React\Socket\Server('[::1]:8080', $loop, array(
   their defaults and effects of changing these may vary depending on your system
   and/or PHP version.
   Passing unknown context options has no effect.
+  The `backlog` context option defaults to `511` unless given explicitly.
   For BC reasons, you can also pass the TCP socket context options as a simple
   array without wrapping this in another array under the `tcp` key.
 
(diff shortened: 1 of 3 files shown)


kelunik:
@clue Care to shed some light why you chose 511 instead of 512?

clue:
> @acasademont 511 is not only used in Node.js and PHP-FPM, but also in Nginx, Apache, Redis, etc... I'd say 511 was determined to be a safe limit on most OSes (except FreeBSD where it's -1). The backlog limit default in PHP was committed 5 years ago by an user, I don't think there was much discussion about it (the limit default itself) before or after adding a backlog stream context option?

_Originally posted by @CharlotteDunois in https://github.com/reactphp/http/issues/354#issuecomment-569425096_

kelunik:
Thanks!

PR #243, opened 2020-08-19
Update PHPUnit configuration schema for PHPUnit 9.3 and replace deprecated at() Mocks

SimonFrings:
PHPUnit 9.3 released a new schema for the `phpunit.xml` configuration file. I had to migrate the file to the new format in order to avoid the warning. PHPUnit Versions older than 9.3 have to use the `phpunit.xml.legacy` configuration file, because the new format is unknown for them.
I also had to update  some tests, because the `at()` matcher deprecated in PHPUnit 9.3 and will be removed in PHPUnit 10.

The code needs some minor changes to run with PHP 8, I will look into this in an upcoming pull request.

For further details concerning this pull request look into [graphp/graphviz #46](https://github.com/graphp/graphviz/pull/46).
This pull request builds on top of [reactphp/socket #239](https://github.com/reactphp/socket/pull/239).

PR #254, opened 2021-02-24
Use GitHub actions for continuous integration (CI)

SimonFrings:
Bye bye Travis CI, you've served us well.
@clue and I are currently working on switching from Travis to GitHub actions in all ReactPHP repositories.

I've built a little workaround for some tests for legacy PHP. When using recursion there is always the risk of reaching a full stack which results in PHP failing hard. Therefore the `max_nesting_level` setting from Xdebug (to prevent reaching this point) was set to 100 for PHP Versions < 5.4 which resulted in failing test for legacy PHP. In 2015 this setting was changed to 256 for PHP Versions > 5.4. For more details look into [Xdebug#1100](https://bugs.xdebug.org/bug_view_page.php?bug_id=00001100)

This code supports running tests on PHP 8.

Builds on top of #243 and reactphp/reactphp#442.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+  pull_request:
+
+jobs:
+  PHPUnit:
+    name: PHPUnit (PHP ${{ matrix.php }} on ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - ubuntu-18.04
```
clue, line 14: I see the other components are running on `ubuntu-20.04`, any reason to use an older distro here?

tests/FunctionalConnectorTest.php
```diff
@@ -40,6 +40,10 @@ public function connectionToTcpServerShouldSucceedWithLocalhost()
      */
     public function testConnectTwiceWithoutHappyEyeBallsOnlySendsSingleDnsQueryDueToLocalDnsCache()
     {
+        if (DIRECTORY_SEPARATOR === '\\' || defined('HHVM_VERSION')) {
+            $this->markTestSkipped();
```
clue, line 44: Why is this skipped?

tests/IntegrationTest.php
```diff
@@ -101,6 +101,10 @@ public function gettingPlaintextStuffFromEncryptedGoogleShouldNotWork()
 
     public function testConnectingFailsIfConnectorUsesInvalidDnsResolverAddress()
     {
+        if (PHP_OS === 'Darwin') {
+            $this->markTestSkipped('Skipped on macOS due to a bug in reactphp/dns (solved in reactphp/dns#171)');
```
clue, line 105: LGTM. Adding a link here only for reference: https://github.com/reactphp/dns/pull/171

.github/workflows/ci.yml
SimonFrings: No reason, just a mistake on my end.

tests/FunctionalConnectorTest.php
SimonFrings: if ((DIRECTORY_SEPARATOR === '\\' && PHP_VERSION_ID < 70000) || defined('HHVM_VERSION')) {
            $this->markTestSkipped('Not supported on windows for PHP versions < 7.0 and legacy HHVM');
        }

clue (approved):
@SimonFrings Thanks for the update, changes LGTM! :shipit:

WyriHaximus (approved):

tests/IntegrationTest.php
clue: For the reference: We might as well update this now that https://github.com/reactphp/dns/pull/171 has been released a month ago. I'm good with this either way. :+1:

jsor (approved):

PR #266, opened 2021-08-11
Improve error messages for failed TCP/IP connections without ext-sockets

clue:
This changeset improves error messages for failed TCP/IP connections when `ext-sockets` is not available on Linux. This means it will now consistently report errno/errstr on most supported platforms such as this:

```bash
$ telnet 192.168.0.1 8080
Trying 192.168.0.1...
telnet: Unable to connect to remote host: Network is unreachable
$ php examples/11-http-client.php 192.168.0.1
RuntimeException: Connection to tcp://192.168.0.1:80 failed: Network is unreachable in …/TcpConnector.php:115
```

Builds on top of #265, #171, reactphp/dns#171, reactphp/dns#172 and others

### DIFF
--- a/src/TcpConnector.php
+++ b/src/TcpConnector.php
@@ -99,14 +99,27 @@ final class TcpConnector implements ConnectorInterface
                 // The following hack looks like the only way to
                 // detect connection refused errors with PHP's stream sockets.
                 if (false === \stream_socket_get_name($stream, true)) {
-                    // actual socket errno and errstr can only be retrieved when ext-sockets is available (see tests)
+                    // If we reach this point, we know the connection is dead, but we don't know the underlying error condition.
                     // @codeCoverageIgnoreStart
                     if (\function_exists('socket_import_stream')) {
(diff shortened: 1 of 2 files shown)


clue:
I've updated this to use the above work-around only when applicable. This is only known to work on Linux and known to not work on Mac and Windows. Using `ext-sockets` to access the `SO_ERROR` remains the preferred way, so most other platforms should be supported as well.

For future reference: This work-around uses "error slippage" which has been known to work for decades, see e.g. https://cr.yp.to/docs/connect.html. The recommended way would be to use a `read(fd, 0)` call, but PHP's `fread($stream, 0)` rejects empty reads and a `fread($stream, 1)` will return `false` without exposing the errno/errstr. Using `fwrite($stream, $data)` is known to work on Linux, but we might be able to find other platforms that could potentially also take advantage of this. Given that this is only used when `ext-sockets` is not available and this changeset already provides value on its own, I would suggest leaving this up for a future PR.

This is ready for review :shipit:

WyriHaximus (approved):

PR #268, opened 2021-08-19
Skip unsupported tests for HHVM

SimonFrings:
The improved error messages in #267 by @clue are not supported under HHVM.


clue (approved):
@SimonFrings Good catch, thanks for looking into this! :shipit:

WyriHaximus (approved):

PR #276, opened 2021-11-29
add tests to internet group as they require a dns

remicollet:
To avoid having to filter these test during build without internet connection, without dns


clue (commented):
@remicollet Have you tried verifying these tests do in fact require an internet connection? Both tests use a mocked DNS resolver, so they should not actually access the internet. I've just tested this locally and running with `--exclude-group internet` works just fine with no connectivity.

remicollet:
Indeed, seems an old leftover of some old versions

PR #283, opened 2022-01-13
Use default loop in testsuite and close server and connection leftovers

SimonFrings:
I noticed that this testsuite still uses the old eventloop implementation. After I changed everything to the new default loop some tests got stuck in execution, because other tests didn't close their servers/connections.

Builds on top of #260 and  clue/reactphp-block#60.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "react/stream": "^1.2"
     },
     "require-dev": {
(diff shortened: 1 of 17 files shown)


clue (approved):
@SimonFrings Thanks for looking into this, changes LGTM! :shipit:

WyriHaximus (approved):

PR #288, opened 2022-03-22
Fix optional test for `ENETUNREACH` error condition

clue:
This minor changeset fixes the optional test for the `ENETUNREACH` error condition. This test is skipped on many platforms and as such also wasn't executed/spotted on GitHub Actions. It looks like this minor test issue has been introduced via #283. Originally introduced via #265. Spotted while working on https://github.com/reactphp/reactphp/pull/454


SimonFrings (approved):

clue:
For the reference: Given that this test is skipped on most platforms anyway, I've reduced the time it takes to skip from ~20s to ~2s. The total test time is now down from ~23s to ~5s.

WyriHaximus (approved):

PR #304, opened 2023-04-13
Improve errno detection for failed connections without ext-sockets

clue:
This changeset improves errno detection for failed connections without ext-sockets. Reporting the errstr already works on most platforms even when `ext-sockets` is not installed, consistently reporting the correct errno previously required `ext-sockets`. With these changes applied, we now also take the errno constants from `ext-pcntl` and `ext-posix` into account. This is particularly useful for lightweight container images (Docker) that by default come with `ext-sockets` and `ext-pcntl` disabled but `ext-posix` enabled. All extensions remain entirely optional.

Link dump:
* https://man7.org/linux/man-pages/man3/errno.3.html
* https://man7.org/linux/man-pages/man3/strerror.3.html
* https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/errno.h
* https://learn.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=msvc-170

Builds on top of #271, #267, #266 and others

### DIFF
--- a/src/Connector.php
+++ b/src/Connector.php
@@ -170,7 +170,7 @@ final class Connector implements ConnectorInterface
         if (!isset($this->connectors[$scheme])) {
             return \React\Promise\reject(new \RuntimeException(
                 'No connector available for URI scheme "' . $scheme . '" (EINVAL)',
-                \defined('SOCKET_EINVAL') ? \SOCKET_EINVAL : 22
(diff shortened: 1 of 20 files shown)


SimonFrings (approved):
:+1:

WyriHaximus (approved):

PR #323, opened 2025-04-29
Update test suite to use PCOV to avoid segfault with Xdebug 3.4.2

clue:
This changeset updates the test suite to use PCOV instead of Xdebug on PHP 8+ to avoid a segfault with Xdebug 3.4.2 on PHP 8.1+ (https://bugs.xdebug.org/view.php?id=2332). 

Unfortunately, it doesn't look like the upstream bug in Xdebug is an easy fix nor like it is going to be fixed any time soon, so for the time being I'd propose switching to PCOV to unblock our development. PCOV is pretty much a drop-in replacement on newer PHP versions (but noticeably faster), and test coverage should not be affected at all. Once merged, we should apply similar changes to our other components and other supported branches.

Builds on top of https://github.com/clue/framework-x/pull/278 and #321 
Supersedes / closes #322

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          coverage: xdebug
+          coverage: ${{ matrix.php < 8.0 && 'xdebug' || 'pcov' }}
           ini-file: development
       - run: composer install
       - run: vendor/bin/phpunit --coverage-text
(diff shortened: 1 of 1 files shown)


WyriHaximus (approved):
👍

clue:
For the reference: Xdebug 3.4.5 has been released last week (https://github.com/xdebug/xdebug/releases/tag/3.4.5) with the above bug fixed, so the segfault should no longer be an issue.

We've briefly discussed reverting this internally and figured this would introduce unneeded noise at the moment, so we'll likely stick with a mix of Xdebug and PCOV for any projects migrated in the meantime. There is a chance either may break in the future, so I think it's good we have both at our disposal in case this happens.

PR #329, opened 2025-10-24
Catch fatal exception in stream_socket_client

### LINKED ISSUE #328: stream_socket_client throws fatal error if network is unreachable
https://github.com/reactphp/socket/blob/def0730288b2a321db04f22caf86259ab2975cc7/src/TcpConnector.php#L91

This should be handled more gracefully by returning with a rejected promise if it fails to connect.

valzargaming:
Closes #328


clue:
@valzargaming Thanks, noted your PR here. Let's first continue the discussion in #328 to get a better understanding what we're trying to fix here.

